PR #54, opened 2022-05-20
Hamidra/pool memberships

hamidra:
This will add pool membership subscriptions for the active account to the pool context.
I think we can use the membership to specify if the active account is participating in any pools.


rossbulat (approved):

rossbulat (approved):
Nice addition, looks good!

PR #95, opened 2022-06-03
Unbond and Leave from pool

hamidra:
This should add complete pool functionality to pool page.
 - [x] claim
 - [x] Withdraw
 - [x]  Leave pool
 - [x] Change pool state. (destroy, lock, unlock)
 
 This does not include role management.
for role managment the admin (owner or state-toggler) should be able to assign the roles to other accounts. For that we need to have a design for the edit page, to make sure proper safety precautions are put in that form to prevent the admin from transferring roles to wrong addresses and lock themselves out.
 We can consider that as an advanced feature to be added later. Currently if anyone wants to change the roles they need to submit the tx manually or through polkadot.js app.


hamidra:
Screen Shot 2022-06-06 at 6 24 04 PM

rossbulat (approved):

rossbulat:
Excellent progress, I made a few amendments, mostly just general fixing.

Still to do for Pools post this PR:
 - extrinsic to stop nominating active nominations
 - ability edit pool roles
 - add locked status to “pool status”
 - pool destroying state in Status, with an “unbond all users” extrinsic.

in any case, I'll continue testing this PR post merge with KSM that I'm about to set up! Would be good to have some payout claims to test.

PR #97, opened 2022-06-04
renaming sections

TingALin:
Adresses #92 

- Renaming all the SectionWrapper to CardWrapper, and SectionsWrapper to CardsWrapper, also SectionHeaderWrapper to CardHeaderWrapper
- Keeping the Sections file name under the library/Assistant but edited the SectionWrapper and same to Modals/connectaccounts /undatebond /unlockchunks


rossbulat (approved):
Looks good!

PR #136, opened 2022-06-24
Add edit button to pool roles.

hamidra:
This will make the nominator and stateToggler roles editable.
There are some concerns about making the root editable since the pool owner might lock themselves out by setting the wrong address so these changes won't make the root editable.

Screen Shot 2022-06-24 at 4 35 56 PM

Screen Shot 2022-06-24 at 4 52 37 PM


rossbulat (approved):
A very elegant and easy way to update pool roles! Great use of the modal components that keep UI inline with UpdateController.

PR #215, opened 2022-08-16
add Generic-Chain

Generic-Chain:
add Generic-Chain Validators to the staking dashboard


rossbulat:
Added on master branch. Thank you!

PR #285, opened 2022-10-22
Feature/setup jest

l0r1s:
- Added eslint-plugin-jest and configured jest/globals as env
- Configured *jest.transformIgnorePatterns* in package.json to fix non-transpiled packages

The code used to test the configuration is working as expected is the following:

```tsx
import { renderHook } from '@testing-library/react';
import { DefaultNetwork } from 'consts';
import { APIProvider, useApi } from '../Api';

describe('APIProvider', () => {
  it('should works', () => {
    localStorage.setItem('network', DefaultNetwork);

    const { result } = renderHook(() => useApi(), { wrapper: APIProvider });

    expect(result).toBeDefined();
  });
});
```

Before configuring jest.transformIgnorePatterns, the error was this one:
Screenshot from 2022-10-22 11-42-50

After the introduction of the configuration option, there is no more error, jest is able to run babel on needed node_modules:
Screenshot from 2022-10-22 11-43-30

Any comment or suggestion is welcome


rossbulat:
Looks good to me. I have asked @wirednkod for his thoughts on Jest. I haven't done too much reading on it myself yet - are there any major configs we should add to this?

l0r1s:
Jest is mainly pre-configured by react-scripts, you can get the full configuration using ```yarn test --showConfig```

Here we ate just overriding one option

wirednkod:
Thamk you @rossbulat , @l0r1s . I have a question though - why not `ts-jest`?  
I mean - the project is in Typescript and it would make more sense to me. Also, I do not see any `jest.config` in the PR that will "take care" of the transform (e.g. using babel-jest).

wirednkod (dismissed):
@l0r1s just to clarify - I understand that CRA supports out of the box the `testing-library` - and thus it is maybe better to go down that path at the moment. Im just wondering (and this we have discussed before - @rossbulat ), if we want to fill up tests that are tight with CRA. Im more into 'not-using-CRA' projects, and thus I would prefer separate test tools if possible - but this is just my personal opinion. I would leave the final decision to Ross on the roadmap.

Having said that, youor PR looks good to me - I would just add the test you presented above, into the PR, for having a PoC that `test` works ;) 

Thank you again!

l0r1s:
Hi @wirednkod,

You are right, we could have used ts-jest, but react-scripts was already configured for testing, so It was the easiest path to take.

I understand your point of view, now having all the information about not getting coupled with react-scripts.

react-scripts is using Jest with babel-jest under the hood, and there is a function to generate the config here:
https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/scripts/utils/createJestConfig.js

The only pitfall I see in using babel-jest (through react-scripts) is the type checking which is not robust compared to ts-jest, as highlighted in the ts-jest documentation:
https://kulshekhar.github.io/ts-jest/docs/babel7-or-ts/

There is no problem for me to use ts-jest and I can update the PR accordingly (adding the PoC test to be able to check on your side), I don't know what you are using on other TS projects inside Parity but it could be useful to uniformize?

wirednkod:
> You are right, we could have used ts-jest, but react-scripts was already configured for testing, so It was the easiest path to take. 
> I understand your point of view, now having all the information about not getting coupled with react-scripts.

My suggestion would be to use in addition `ts-jest` especially for type checking - and also have at least 1 test implemented (as per your example) - just for testing purposes. I have run your example and it works just fine (thus I approved the PR).

> There is no problem for me to use ts-jest and I can update the PR accordingly (adding the PoC test to be able to check on your side), I don't know what you are using on other TS projects inside Parity but it could be useful to uniformize?

In other projects I have seen, we use mainly `ts-jest` but agan - i'm not strongly opinionated about. I just think its a good add-on on top. I think even if project is decoupled from CRA, `testing-library` can still be used.

Its a great kick-off PR for adding tests in the project. Thank you for the contribution

rossbulat:
Thanks @l0r1s and @wirednkod!

Let's move forward with the eventual goal of decoupling fro CRA.
- Let's use `ts-jest` 
- Please do add the test case you ran in the screenshot.

Then we'll merge and open the doors for more tests :)

rossbulat:
Could we also remove the `testing-library` dependencies how since we are going with jest?

Screenshot 2022-10-27 at 13 17 36

wirednkod:
> Could we also remove the `testing-library` dependencies now since we are going with jest?
> 
> Screenshot 2022-10-27 at 13 17 36

I think we should not... I think testing-library is the best way to test react, CRA or not

rossbulat:
Closing this as Jest is now introduced via #521. Thank you for kickstarting with this issue.

PR #527, opened 2023-01-03
Change unbond duration from eras to time left

TingALin:
This PR links to issue #520 

- Create new hook, useErasToTimeleft, where you can pass a number of eras into the hook and get the corresponding duration in seconds back.
- library/Hooks/useTimeLeft.tsx: Evolve timeleftAsString(): string to timeLeftAsString(to?: Date):string. Handle the current era (toRef.current) if no to is provided as an argument.
- Have an option to either do “full” format (hours, minutes) or not (hrs, mins). Full can be a Boolean argument that defaults to false
- Get rid of “0 hours” /“0 minutes”if hrs/mins is zero


src/modals/UnlockChunks/Overview.tsx
```diff
@@ -25,6 +28,14 @@ export const Overview = forwardRef(
     const { isFastUnstaking } = useUnstaking();
     const { t } = useTranslation('modals');
 
+    const { timeleft: eraTimeLeft } = useEraTimeLeft();
+    const { fromNow, timeleftAsString } = useTimeLeft({
+      refreshInterval: 60,
+      refreshCallback: () => eraTimeLeft,
+    });
+    const { durationSeconds } = useErasToTimeleft(bondDuration);
```
rossbulat, line 36: When I tested the modal the text flickers between `12 hours` and `11 hours 59 mins` every so often. 

This value does not need to be a live, it is always 12 hours that needs to be statically converted from eras to time.

src/library/Hooks/useTimeLeft/index.tsx
```diff
@@ -184,21 +184,31 @@ export const useTimeLeft = ({
   }, []);
 
   // format the duration as a string.
-  const timeleftAsString = (() => {
-    const { days, hours, minutes, seconds } = getDuration(toRef.current);
+  const timeleftAsString = (toDate?: Date, fullFormat?: boolean) => {
```
rossbulat, line 187: change `fullFormat` to just `full`.

src/library/Hooks/useErasToTimeleft/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+  // store the duration of an era in block numbers.
+  const eraDurationBlocks = epochDuration * sessionsPerEra;
+
+  // estimate the duration of the era in seconds
+  const eraDuration = eraDurationBlocks * expectedBlockTime * 0.001;
+
+  const durationSeconds = eras * eraDuration;
+
+  return { durationSeconds };
```
rossbulat, line 18: Do not make this hook calculate the timeleft immediately after initialisation. 

Instead, let it provide a function that can be imported to a component and called when needed.

```suggestion

const erasToTimeLeft = (eras: number) => {
  const { consts } = useApi();
  const { epochDuration, expectedBlockTime, sessionsPerEra } = consts;
  // store the duration of an era in block numbers.
  const eraDurationBlocks = epochDuration * sessionsPerEra;
  // estimate the duration of the era in seconds
  const eraDuration = eraDurationBlocks * expectedBlockTime * 0.001;
  const durationSeconds = eras * eraDuration;
  return durationSeconds;
  }
  
  return {
     erasToTimeLeft
  };
```

src/modals/UnlockChunks/Overview.tsx
```diff
@@ -25,6 +28,14 @@ export const Overview = forwardRef(
     const { isFastUnstaking } = useUnstaking();
     const { t } = useTranslation('modals');
 
+    const { timeleft: eraTimeLeft } = useEraTimeLeft();
```
rossbulat, line 31: This is not needed. We are not working with a live era count down.

src/modals/UnlockChunks/Overview.tsx
```diff
@@ -25,6 +28,14 @@ export const Overview = forwardRef(
...
+    const { fromNow, timeleftAsString } = useTimeLeft({
+      refreshInterval: 60,
+      refreshCallback: () => eraTimeLeft,
```
rossbulat, line 34: Bit of a detour but can these be optional? Default `refreshInterval` to 60 and only call `refreshCallback` if it is provided.

rossbulat (commented):
On the right track but some changes to be made. 

Noticed a bug where the time flickers every once in a while, and you are using `useEraTimeLeft` in the modal unnecessarily.

src/library/Hooks/useErasToTimeleft/index.tsx
rossbulat: I don't understand the question.

src/library/Hooks/useTimeLeft/types.ts
```diff
@@ -28,6 +28,6 @@ export interface TimeLeftAll {
 }
 
 export interface TimeleftHookProps {
-  refreshCallback: () => number;
+  refreshCallback?: () => number;
```
rossbulat, line 31: remove ?

rossbulat:
Screenshot 2023-01-04 at 15 23 17

Can you change this text to: "Unlocks can be withdrawn after 12 hours."

src/library/Hooks/useTimeLeft/index.tsx
```diff
@@ -117,58 +114,63 @@ export const useTimeLeft = ({
...
-    //
-    // either handle regular timeleft update, or refresh `to` via `refreshCallback` every
-    // `refreshInterval` seconds.
-    const handleRefresh = () => {
-      if (r !== 0) {
-        setTimeleft(getTimeleft());
-      } else {
-        if (refreshCallback) {
-          setStateWithRef(fromNow(refreshCallback()), setTo, toRef);
+  if (TimeleftHook) {
```
rossbulat, line 117: local variables should not be capitalised. Only globals should be.

src/library/Hooks/useTimeLeft/index.tsx
```diff
@@ -117,58 +114,63 @@ export const useTimeLeft = ({
...
-        const interval = setInterval(() => {
-          if (inLastHour()) {
-            clearInterval(minIntervalRef.current);
-            setStateWithRef(undefined, setMinInterval, minIntervalRef);
-          }
-          r = Math.max(0, r - 60);
-          handleRefresh();
-        }, 60000);
-        setStateWithRef(interval, setMinInterval, minIntervalRef);
+          r = TimeleftHook.refreshInterval;
```
rossbulat, line 139: This does not fall back to a default value.
`r  = timeleftHook?.refreshInterval || 60;`

can introduce defaults here too.

src/library/Hooks/useTimeLeft/index.tsx
rossbulat, line 117: useEffects should **Never** be scoped or wrapped in anything. They need to be top level

src/library/Hooks/useTimeLeft/index.tsx
```diff
@@ -14,10 +14,7 @@ import {
   TimeLeftRaw,
 } from './types';
 
-export const useTimeLeft = ({
-  refreshCallback,
-  refreshInterval,
-}: TimeleftHookProps) => {
+export const useTimeLeft = (TimeleftHook?: TimeleftHookProps) => {
```
rossbulat: `TmeleftHook` is not a hook, they can be treated as props to the hook. Therefore this should just be called `props`.

src/library/Hooks/useTimeLeft/index.tsx
```diff
@@ -184,21 +186,31 @@ export const useTimeLeft = ({
   }, []);
 
   // format the duration as a string.
-  const timeleftAsString = (() => {
-    const { days, hours, minutes, seconds } = getDuration(toRef.current);
+  const timeleftAsString = (toDate?: Date, full?: boolean) => {
```
rossbulat, line 188: `toDate` can just be called `to`

src/library/Hooks/useTimeLeft/index.tsx
```diff
@@ -184,21 +186,31 @@ export const useTimeLeft = ({
...
+    const { days, hours, minutes, seconds } = toDate
+      ? getDuration(toDate)
+      : getDuration(toRef.current);
```
rossbulat, line 191:
```suggestion
    const { days, hours, minutes, seconds } = getDuration(toDate || toRef.current);
```

src/library/Hooks/useTimeLeft/index.tsx
```diff
@@ -184,21 +186,31 @@ export const useTimeLeft = ({
...
+    if (hours > 0) {
+      str += full
+        ? `${hours} ${t('time.hour', { count: hours, ns: 'base' })} `
+        : `${hours} ${t('time.hr', { count: hours, ns: 'base' })} `;
+    }
+    if (minutes > 0) {
+      str += full
+        ? `${minutes} ${t('time.minute', { count: minutes, ns: 'base' })}`
+        : `${minutes} ${t('time.min', { count: minutes, ns: 'base' })}`;
+    }
```
rossbulat, line 206:
```suggestion
    let tHour = full ? 'time.hour' : 'time.hr';
    let tMinute = full ? 'time.minute' : 'time.min';
    
    if (hours > 0) {
      str += `${hours} ${t(tHour, { count: hours, ns: 'base' })} `
    }
    if (minutes > 0) {
      str +=
        `${minutes} ${t(tMinute, { count: minutes, ns: 'base' })}`
    }
```

src/library/StatBoxList/Timeleft.tsx
```diff
@@ -6,7 +6,7 @@ import { StatPie } from 'library/Graphs/StatBoxPie';
 import { OpenHelpIcon } from 'library/OpenHelpIcon';
 import { StatBox } from './Item';
 import { TimeleftProps } from './types';
-import { TineleftWrapper } from './Wrapper';
+import { TimeleftWrapper } from './Wrapper';
```
rossbulat, line 9: Let's make sure `t` is capitalised in timeleft. Rename to `TimeLeftWrapper`.

src/library/Hooks/useErasToTimeleft/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { useApi } from 'contexts/Api';
+
+export const useErasToTimeleft = () => {
+  const { consts } = useApi();
+  const { epochDuration, expectedBlockTime, sessionsPerEra } = consts;
+
+  // converts a number of eras to timeleft in seconds.
+  const erasToTimeLeft = (eras: number) => {
```
rossbulat, line 11: Rename this function to `getErasToTimeLeft`.

When we are getting something (as opposed to setting something), we can use the `get` prefix.

TingALin:
> Unlocks can be withdrawn after 12 hours.

done

rossbulat (approved):

rossbulat (approved):

PR #539, opened 2023-01-06
Introduce URL variables

TingALin:
This links to issue #531


public/favicons/polkadot/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<browserconfig>
+    <msapplication>
+        <tile>
+            <square150x150logo src="/mstile-150x150.png"/>
+            <TileColor>#da532c</TileColor>
```
rossbulat, line 6: Why is this orange?

public/index.html
```diff
@@ -5,10 +5,10 @@
...
-  <link rel="apple-touch-icon" sizes="180x180" href="%PUBLIC_URL%/favicons/apple-touch-icon.png">
-  <link rel="icon" type="image/png" sizes="32x32" href="%PUBLIC_URL%/favicons/favicon-32x32.png">
-  <link rel="icon" type="image/png" sizes="16x16" href="%PUBLIC_URL%/favicons/favicon-16x16.png">
-  <link rel="mask-icon" href="%PUBLIC_URL%/favicons/safari-pinned-tab.svg" color="#e6007a">
+  <link rel="apple-touch-icon" sizes="180x180" href="%PUBLIC_URL%/favicons/polkadot/apple-touch-icon.png">
+  <link rel="icon" type="image/png" sizes="32x32" href="%PUBLIC_URL%/favicons/polkadot/favicon-32x32.png">
+  <link rel="icon" type="image/png" sizes="16x16" href="%PUBLIC_URL%/favicons/polkadot/favicon-16x16.png">
+  <link rel="mask-icon" href="%PUBLIC_URL%/favicons/polkadot/safari-pinned-tab.svg" color="#e6007a">
   <meta name="msapplication-TileColor" content="#e6007a">
   <meta name="theme-color" content="#fff">
```
rossbulat, line 13: I don't think these 2 meta tags are updated when the network is switched?

src/Router.tsx
```diff
@@ -40,6 +42,68 @@ export const RouterInner = () => {
   const { sideMenuOpen, sideMenuMinimised, setContainerRefs } = useUi();
   const { t } = useTranslation('base');
 
+  const [urlNetworkExists, setUrlNetworkExists] = useState<boolean>(false);
+  const [urlNetwork, setUrlNetwork] = useState<any>('');
+  const [urlLngExists, setUrlLngExists] = useState<boolean>(false);
+  const [urlLng, setUrlLng] = useState<any>('');
```
rossbulat, line 48: None of these state items are needed. The active `network` and `switchNetwork` function are stored in the `API` context, and the active locale is  managed by `i18n.resolvedLanguage` / `i18n.changeLanguage()`. 

Use these directly. If the URL vars do not exist, nothing needs to be done.

src/Router.tsx
```diff
@@ -40,6 +42,68 @@ export const RouterInner = () => {
...
+      (urlNetworkSource === 'polkadot' || 'kusama' || 'westend')
+    ) {
+      setUrlNetworkExists(true);
+      setUrlNetwork(urlNetworkSource);
+    }
+    const _network = urlNetworkExists
+      ? urlNetwork
+      : network
+      ? network.name.toLowerCase()
+      : DefaultNetwork;
```
rossbulat, line 64: Something to be careful of is that `DefaultNetwork` is all lower case, whereas `NETWORKS[x].name is capitalised. We should resolve this and stick to one convention.

rossbulat (changes requested):
A start, but lots of tidy up to do. Will do some testing after the comments are addressed.

public/favicons/polkadot/browserconfig.xml
TingALin: Just realized this is the suggested color for the Windows Metro UI and I didn't edit it,  now I know that 3 favicons are using the same tilecolor.
And for the kusama and westend icon, I chose the inline ones from the img folder, so let me know if i should use other's images for the favicons.

public/index.html
TingALin: I didn't see there's any obvious differences between the background color change even they are using different color number, I will edit it

src/Router.tsx
```diff
@@ -35,10 +40,27 @@ import {
...
+    const urlNetworkSource = extractUrlValue('n');
+    if (
+      Object.values(NETWORKS).find(
+        (n: any) => n.name.toLowerCase() === urlNetworkSource
+      )
+    ) {
+      switchNetwork(urlNetworkSource as NetworkName, true);
+    }
+
+    const urlLngSource = extractUrlValue('l');
```
rossbulat, line 58: `lngFromUrl`

rossbulat:
Please also use `t` to translate the `TitleDefault`. `TitleDefault` can be discontinued.

Screenshot 2023-01-07 at 22 52 49

src/library/Hooks/useUrlVars/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { changeLanguage } from 'locale/utils';
+import { useTranslation } from 'react-i18next';
+import { NetworkName } from 'types';
+import { extractUrlValue } from 'Utils';
+
+export const useUrlVars = () => {
+  const { i18n } = useTranslation();
+  // get url variables
+  const networkFromUrl = extractUrlValue('n');
+  const lngFromUrl = extractUrlValue('l');
```
rossbulat, line 16: These can be inside `initialise()` at the top. They are a part of `initialise`- although there is no harm in calling them here (app will still work) let's keep all initialisation logic together.

rossbulat:
@TingALin Can you amend the url when a network is switched? e.g. if `n=kusama` is in the URL, and a user switches to polkadot, change the URL to be `n=polkadot`.

PR #648, opened 2023-02-08
Convert API consts from number to BigNumber

TingALin:
This PR links to issue #637 

Notes: 
- Make no differences to make the number as number type, but just keep the 0 in BigNumber format for the consistency
- Any type can be in the arithmetic operation with BigNumber with no errors


src/consts.ts
```diff
@@ -39,13 +40,13 @@ export const PluginsList = ['subscan', 'binance_spot', 'tips'];
...
-export const FallbackBondingDuration = 28;
-export const FallbackSessionsPerEra = 6;
-export const FallbackNominatorRewardedPerValidator = 256;
-export const FallbackMaxElectingVoters = 22500;
-export const FallbackExpectedBlockTime = 6000;
-export const FallbackEpochDuration = 2400;
+export const FallbackMaxNominations = new BigNumber(16);
+export const FallbackBondingDuration = new BigNumber(28);
+export const FallbackSessionsPerEra = new BigNumber(6);
+export const FallbackNominatorRewardedPerValidator = new BigNumber(256);
```
rossbulat, line 46: This is now 512

src/contexts/FastUnstake/index.tsx
```diff
@@ -132,7 +132,7 @@ export const FastUnstakeProvider = ({
         // if localMeta existed, start checking from the next era.
         const eraTocheck = localMeta
           ? localMeta.checked[localMeta.checked.length - 1] - 1
-          : activeEra.index;
+          : activeEra.index.toNumber();
```
rossbulat, line 135: This should ideally be BigNumber, as we are converting all era indexes to BigNumber.

src/contexts/FastUnstake/index.tsx
```diff
@@ -219,7 +219,7 @@ export const FastUnstakeProvider = ({
         subscribeToFastUnstakeQueue();
       } else {
         // continue checking the next era.
-        checkEra(currentEra - 1);
+        checkEra(new BigNumber(currentEra - 1));
```
rossbulat, line 222: `currentEra` is a big number so this should be `currentEra.minus(1)`

src/contexts/Validators/index.tsx
```diff
@@ -505,30 +505,33 @@ export const ValidatorsProvider = ({
...
+          others = others.sort((a: AnyApi, b: AnyApi) =>
+            a.value.minus(b.value)
+          );
 
           // get the lowest reward stake of the validator, which is
           // the largest index - `maxNominatorRewardedPerValidator`,
           // or the first index if does not exist.
-          const lowestRewardIndex = Math.max(
-            others.length - maxNominatorRewardedPerValidator,
+          const lowestRewardIndex = BigNumber.max(
```
rossbulat, line 525: I actually think this index should just be a `number`, we are just using it as an array index right now, which is a `number` anyway.

src/contexts/Validators/index.tsx
```diff
@@ -505,30 +505,33 @@ export const ValidatorsProvider = ({
...
             0
           );
 
           const lowestReward =
             others.length > 0
-              ? planckToUnit(
-                  new BigNumber(rmCommas(others[lowestRewardIndex]?.value)),
-                  units
-                )
+              ? planckToUnit(others[lowestRewardIndex.toNumber()]?.value, units)
```
rossbulat, line 534: This line will look better if `lowestRewardIndex` is just a number.

src/library/GenerateNominations/index.tsx
```diff
@@ -194,10 +194,13 @@ export const GenerateNominations = ({
   };
 
   const disabledMaxNominations = () => {
-    return nominations.length >= maxNominations;
+    return maxNominations.isLessThanOrEqualTo(nominations.length);
```
rossbulat, line 197: Why has `maxNominations` been replaced with `nominations.length`? These are 2 different values (hopefully the names are self explanatory. Please fix.

src/library/GenerateNominations/index.tsx
```diff
@@ -194,10 +194,13 @@ export const GenerateNominations = ({
...
   };
   const disabledAddFavorites = () => {
-    return !favoritesList?.length || nominations.length >= maxNominations;
+    return (
+      !favoritesList?.length ||
+      maxNominations.isLessThanOrEqualTo(nominations.length)
```
rossbulat, line 202: This is different logic to what was there before. You've switched the numbers around so it should be `isLessThan`.

src/library/Hooks/useValidatorFilters/index.tsx
```diff
@@ -78,7 +78,7 @@ export const useValidatorFilters = () => {
         continue;
       }
       const totalNominations = stake[addressBatchIndex].total_nominations ?? 0;
-      if (totalNominations < maxNominatorRewardedPerValidator) {
+      if (maxNominatorRewardedPerValidator.isGreaterThan(totalNominations)) {
```
rossbulat, line 81: Again, the numbers have been switched so this should be `isGreaterThanOrEqualTo`. But since these are both big numbers now, it can be how it was before with the same operator.

src/library/ListItem/Labels/Oversubscribed.tsx
```diff
@@ -46,7 +46,10 @@ export const Oversubscribed = ({
   const lowestReward = eraStakers?.lowestReward ?? 0;
 
   const displayOversubscribed =
-    synced.stake && totalNominations >= consts.maxNominatorRewardedPerValidator;
+    synced.stake &&
+    consts.maxNominatorRewardedPerValidator.isLessThanOrEqualTo(
```
rossbulat, line 50: Same issue, logic is different. Should be `isLessThan`, and since the numbers are now both big numbers it does not need to change.

src/modals/NominateFromFavorites/index.tsx
```diff
@@ -88,7 +88,7 @@ export const NominateFromFavorites = () => {
   };
 
   const totalAfterSelection = nominations.length + selectedFavorites.length;
-  const overMaxNominations = totalAfterSelection > maxNominations;
+  const overMaxNominations = maxNominations.isLessThan(totalAfterSelection);
```
rossbulat, line 91: Different logic again.

src/modals/SelectFavorites/index.tsx
```diff
@@ -60,7 +60,7 @@ export const SelectFavorites = () => {
   };
 
   const totalAfterSelection = nominations.length + selectedFavorites.length;
-  const overMaxNominations = totalAfterSelection > maxNominations;
+  const overMaxNominations = maxNominations.isLessThan(totalAfterSelection);
```
rossbulat, line 63: same error as above.

src/pages/Nominate/Active/UnstakePrompts.tsx
```diff
@@ -62,7 +62,7 @@ export const UnstakePrompts = () => {
                   <ButtonPrimary
                     marginRight
                     iconLeft={faBolt}
-                    text={getFastUnstakeText()}
+                    text={getFastUnstakeText() as string}
```
rossbulat, line 65: why is `as string()` here?

src/pages/Pools/Home/MembersList/index.tsx
```diff
@@ -86,7 +86,7 @@ export const MembersListInner = ({
 
   // configure list when network is ready to fetch
   useEffect(() => {
-    if (isReady && activeEra.index !== 0 && fetched === 'unsynced') {
+    if (isReady && !activeEra.index.isZero() && fetched === 'unsynced') {
```
rossbulat, line 89: `!activeEra.index.isZero()` is a bit hard to read and is being used in a few places. I think a new utility, `isNotZero()`, would be easier to read.

rossbulat (changes requested):
In addition to the comments:

- BigNumber is not working when being passed to a service worker. Convert to string when passing between worker and main thread.
- There is a fast unstake api request error, I am guessing a BigNumber has been passed as an api argument instead of a string.

There are also quite a few math errors pertaining to the "is" methods that need to be fixed.

src/library/GenerateNominations/index.tsx
TingALin: I'm trying to understand the difference, as the results should be the same, i switched the two values positions

src/library/GenerateNominations/index.tsx
TingALin: why miss the equalTo?

src/library/GenerateNominations/index.tsx
rossbulat: The results are not the same because the = will no longer apply when you switch the values around, a >= b switched is b < a.

src/library/GenerateNominations/index.tsx
rossbulat: because a >= b != b <= a.

a >= b == b < a

TingALin: Sorry for the rookie mistakes and thank you for the patience

src/contexts/FastUnstake/index.tsx
```diff
@@ -180,7 +180,7 @@ export const FastUnstakeProvider = ({
 
       // ensure checked eras are in order highest first.
       const checked = metaRef.current.checked
-        .concat(currentEra)
+        .concat(new BigNumber(currentEra).toNumber())
```
rossbulat, line 183:
```suggestion
        .concat(Number(currentEra))
```

src/contexts/FastUnstake/index.tsx
```diff
@@ -180,7 +180,7 @@ export const FastUnstakeProvider = ({
...
         .sort((a: number, b: number) => b - a);
 
       if (!metaRef.current.checked.includes(currentEra)) {
```
rossbulat, line 186: `currentEra` is still a string. Convert to `Number`:
```suggestion
      if (!metaRef.current.checked.includes(Number(currentEra))) {
```

rossbulat (changes requested):
A couple of things to change, please let me know when the amendments have been made and I will do another - probably final - review.

src/contexts/Network/types.ts
```diff
@@ -16,6 +16,6 @@ export interface NetworkMetrics {
 }
 
 export interface ActiveEra {
-  index: number;
-  start: number;
+  index: BigNumber;
+  start: BigNumber;
```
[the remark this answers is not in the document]
TingALin: What if both of the timestamps got too big in the future? activeEra.start should be bigger than index, so i think you mean index should be number?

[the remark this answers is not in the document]
rossbulat: No, i mean index. Let's just keep it BigNumber for consistency.

rossbulat (approved):
Just requested a testing flow run and then good to merge.

rossbulat (approved):

PR #713, opened 2023-03-01
Remove downshift dependency for switching controller

TingALin:
Replace downshift component with selected state
Rename the css wrapper links to downshift


rossbulat:
Is this ready to be reviewed?

rossbulat:
I can select accounts that do not have enough DOT.

screen

Can you disable these buttons when they are < ED. There should be a class already defined that changes the opacity when these are disabled. (the `outline` property should be `none` on these buttons - should prevent the blue border in the screenshot.

TingALin:
> I can select accounts that do not have enough DOT.
> 
> screen
> 
> Can you disable these buttons when they are < ED. There should be a class already defined that changes the opacity when these are disabled. (the `outline` property should be `none` on these buttons - should prevent the blue border in the screenshot.

fixed

src/library/Form/AccountDropdown/Wrappers.ts
```diff
@@ -96,6 +96,7 @@ export const StyledDropdown = styled.div<any>`
       &.inactive {
         opacity: 0.5;
         cursor: default;
+        border: none;
```
rossbulat, line 99: is this needed?

src/library/Form/AccountDropdown/index.tsx
```diff
@@ -73,7 +73,9 @@ const DropdownItem = ({ item, onChange, selected }: any) => {
   const { unit } = useApi().network;
   return (
     <button
-      className={selected ? 'selected' : undefined}
+      className={`${selected ? 'selected' : undefined} ${
+        item.active ? '' : ' inactive'
```
rossbulat, line 77:
```suggestion
!item.active ? `inactive` : ``
```

src/library/Form/AccountDropdown/Wrappers.ts
TingALin: yes, as outline:none is not working, this one does

rossbulat: the button should just be disabled if not active, which will then make it not selectable.

rossbulat:
All okay, good bye downshift.

PR #874, opened 2023-04-10
Migrate to core-ui/PageRow / Page

TingALin:
Addresses #854 

- Replace PageWrapper with Page component from the UI library.
- Replace PageRowWrapper with PageRow component from the UI library.


rossbulat (approved):
Amazing, so much simpler.

PR #875, opened 2023-04-10
Migrate to core-ui/Side

TingALin:
Replace SideInterfaceWrapper with Side component from the UI library


rossbulat (approved):
Excellent. We discussed the `localStorageOrDefault` optimisation also that will improve these updates further.

PR #876, opened 2023-04-10
Add Fearless Wallet extension

stefashkaa:
<h1>Description</h1>
<p>Add Fearless Wallet</p>

<h1>Works done</h1>

- [x] Added Fearless Wallet extension


rossbulat:
Hi @stefashkaa , thank you for submitting!

I have installed the fearless web extension but it is not showing within the `injectedWeb3` object. It needs to be exposed in this object with `enable` available:

Screenshot 2023-04-14 at 11 36 36

Would it be lurking around another object?

stefashkaa:
@rossbulat Hello! Did you check it under `http` or `https`? Seems like Fearless wallet supports only `https` for now

rossbulat:
Ah thanks, i was testing over `http`. I will push support to staging and see if it works.

rossbulat:
Added in #928

rossbulat:
Hi @stefashkaa, I have noticed that the fearless extension asks for permission and metadata updates every time an app is visited. I think there is an issue with persisting trusted sites and the latest metadata. Can you confirm?

rossbulat:
Hi @stefashkaa , I am getting the behaviour on version 1.1.0, on Brave browser. Is there a timeout at play that makes you need to re-authorise a website after 10mins or something similar?

rossbulat:
@stefashkaa  are you on Element? If so it would be great to connect on there. My handle is @ross:parity.io.

PR #946, opened 2023-04-25
Replace react-number-easing with @polkadotcloud/odometer

TingALin:
As the title suggests: replacing react-number-easing with react-odometerjs for better animation.


rossbulat (approved):
Definitely an improvement over the previous package. The animations look good and suit the dashboard.

The only issue is that the width of the characters is always the same, so a `1` value will have more whitespace around it than `8` for example. 

Perhaps we can look into adjusting the widths on the `dashboard-ui` side in the future. Again just a nit but resolving this would tempt me to use this more throughout the dashboard.

PR #2674, opened 2025-04-22
feat(locales): add English (UK) localisation

joeljkrb:
Added British English language option to enhance user experience for UK users.

Changes include:
- Created British English (enGB) translation files
- Updated language selection UI to display English (US) and English (UK) options
- Ensured proper spacing and responsive layout in language selector
- Set up automatic detection for en-GB browser settings

This improves the dashboard's accessibility for British users while maintaining the existing US English localisation.


rossbulat (commented):
Instead of duplicating a locale folder specifically for GB, I wonder if there is a way we can just amend the affected locale keys in question by:

- Providing a `-gb` suffix in the `en` locale files only.
- Add a mapping of `locales  -> suffixes`, and add checks to the `validate` script to prevent missing key errors
- Where we expect these localised keys to be used (for any language, not just `en`), call a helper function that either takes the expected prefixed key, or returns the fallback.

`withSuffix('key') => key | keyWithSuffix`

Given:

```
const suffixes: Record<string, Record<string, string>> = {
  en: {
      'en-GB: 'gb' // within "en" locale, the "en-GB" language is supported with the "gb" suffix
    }
}
```

And given `en` locale and `en-GB` language is active, `withSuffix`:
- Gets the currently active locale key `en`, and the active language `en-GB`
- Checks if there is a suffix defined via the `suffixes` mapping
- Checks if the suffixed key exists, and returns it if it does, otherwise returns unsuffixed key

We then just deal with the affected keys without as much duplication.

PR #2874, opened 2025-07-04
fix: allow unbond/unstake for any user with bonded funds, not just active nominators

joeljkrb:
Fixed logic in ManageBond.tsx that previously required users to be actively nominating to access unbond/unstake actions.

Now, any user with bonded nominator funds can unbond, regardless of whether they are currently nominating or have stopped nominating.

This resolves issues for users who:
  - Have stopped nominating but still have bonded funds.
  - Are in a pool but have leftover nominator funds (dual staking cleanup).


rossbulat (approved):
Good catch! Instead of checking if the account is nominating, the script should check whether the user has bonded funds (also required for nominating).

rossbulat (approved):

PR #3263, opened 2026-02-11
feat(locale): Add Korean (한국어) language support

joeljkrb:
Adds Korean (ko) as a new locale option in the Staking Dashboard.

- Create 5 Korean translation files (app, help, modals, pages, tips)
- Register Korean locale in i18n config with date-fns ko format


rossbulat:
Looks good to go @joeljkrb

PR #3292, opened 2026-02-23
feat(ux): Visual progress tracker for unbonding chunks

joeljkrb:
Adds a visual progress bar to each unbonding chunk in the UnlockChunks modal, replacing the text-only display with a clear visual indicator of how far through the unbonding period each chunk is.

- Reusable `ProgressBar` component styled consistently with existing patterns and the transaction submission UI redesign (#3291)
- Per-chunk progress calculation based on current era position relative to bond duration
- Distinct visual states for actively unbonding vs fully unlocked chunks
- Smooth animated transitions on load


packages/app/src/library/ProgressBar/Wrappers.ts
```diff
@@ -0,0 +1,63 @@
...
+export const ProgressFill = styled.div<{
+	$progress: number
+	$status: 'active' | 'complete'
+}>`
+	height: 100%;
+	width: ${({ $progress }) => $progress}%;
+	background: ${({ $status }) =>
+		$status === 'complete'
+			? 'var(--status-success)'
+			: 'linear-gradient(90deg, var(--accent-primary) 0%, var(--accent-primary) 85%, var(--shade-100) 100%)'};
```
rossbulat: `--accent-primary` does not appear to be defined anywhere in the theme; the existing accent scale is `--accent-100` through `--accent-1000`. Because this var is inside the gradient, the whole active fill `background` declaration becomes invalid, so active progress bars can render without the intended fill. Please switch this to an existing accent token.

packages/app/src/modals/UnlockChunks/Wrappers.ts
```diff
@@ -21,34 +34,50 @@ export const ChunkWrapper = styled.div`
   margin-top: 1.25rem;
 
   > div {
-    background: var(--gray-400);
+    background: var(--btn-bg);
```
rossbulat: `--btn-bg` is not defined in the theme. The closest existing token looks like `--btn-bg-secondary`; otherwise this resolves as an invalid value and the chunk card background becomes transparent/default.

packages/app/src/modals/UnlockChunks/Wrappers.ts
```diff
@@ -21,34 +34,50 @@ export const ChunkWrapper = styled.div`
...
+      width: 100%;
+    }
 
-      &:first-child {
-        flex-grow: 1;
+    > .status-line {
+      display: inline-flex;
+      align-items: center;
+      gap: 0.4rem;
+      color: var(--text-secondary);
```
rossbulat: `--text-secondary` is not defined either. That means the status text, and the labels below that use the same token, will fall back to inherited/default color, which is especially risky in dark mode. Please use an existing text/gray token such as `--text-tertiary` or the appropriate `--gray-*` value.

rossbulat (changes requested):
I found a few undefined theme tokens in the new styles. These should be switched to existing theme variables before merge; otherwise the active progress fill and chunk text/background styling will not render as intended.

PR #3390, opened 2026-03-31
feat: add route-based code splitting with React.lazy

joeljkrb:
- Replace static page imports with `React.lazy()` dynamic imports in `config/pages.ts`, so each page is loaded on demand as a separate chunk
- Wrap page rendering in `<Suspense>` in `PageWithTitle`
- Update `PageItem.Entry` type from `FC` to `ComponentType` to accept lazy components


joeljkrb:
Thanks @rossbulat , valid points. A few thoughts:

Route splitting effectively gives us the mode-based split for free: Validators, Operators and PoolsList are all flagged advanced: true, so simple-mode users never navigate to them and never download those chunks. HTTP/2 multiplexing keeps per-route overhead low on the CDN side, and bundle caching covers repeat visits.

To address the loading-flash concern I've swapped the bare <div /> Suspense fallback for the existing Preloader so route transitions don't appear blank.

Strong agree on modals and canvas components being the next target, those are bigger wins since they're rarely opened but currently eager (ManagePool, ChangePoolRoles, Plugins, RewardCalculator, full canvas flows). Happy to open a follow-up PR for that.

PR #3536, opened 2026-06-08
feat(fix): Avoid double-subtracting fee reserve from bond amount

joeljkrb:
The Bond modal's `freeToBond` subtracts `feeReserve` from `totalAdditionalBond` / `transferableBalance`, but both already have `feeReserve` deducted in `getTransferableBalance`. This understated the prefilled and max bond amounts by one fee reserve, and disagreed with the `BondFeedback` component's available figure.

**Changes**
- Remove the duplicate fee-reserve subtraction in modals/Bond/index.tsx
- Drop the now-unused feeReserve


packages/app/src/modals/Bond/index.tsx
```diff
@@ -49,11 +49,13 @@ export const Bond = () => {
 	const isPooling = bondFor === 'pool'
 	const { nominator, transferableBalance } = balances
 
+	// `totalAdditionalBond` and `transferableBalance` already have `feeReserve`
+	// deducted (see `getTransferableBalance`), so it must not be subtracted again
```
rossbulat:
```suggestion
	// `totalAdditionalBond` and `transferableBalance` already have `feeReserve`
	// deducted.
```

PR #3551, opened 2026-06-09
feat(ux): Keyboard dismissal, focus management and dialog roles for overlays

joeljkrb:
## Summary

Adds keyboard dismissal for overlays:

- Pressing `Escape` closes the active canvas.
- If no canvas is active, it closes the active modal.
- Prompts remain open and prevent underlying overlays from closing.

The handling is centralized in `OverlayProvider` with no additional hooks, components, dependencies, or accessibility behavior changes.

## Testing

- `pnpm check` passes.


packages/app/src/library/OverlayKeyboardControls/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+// Closes the topmost open overlay when the Escape key is pressed. Centralised
+// here because modals and canvases live in the overlay context while prompts
+// live in a separate context, and prompts render on top of modals/canvases.
+export const OverlayKeyboardControls = () => {
+	const {
+		modal: { status: modalStatus, closeModal },
+		canvas: { status: canvasStatus, closeCanvas },
+	} = useOverlay()
+	const { status: promptStatus, closePrompt, closeOnOutsideClick } = usePrompt()
```
rossbulat: I'm thinking the prompt components could live in `ui-overlay`, then all overlay form factors will all live in that one package.

Perhaps then we can introduce a more intelligent layering system where layer indexes are determined at runtime.

rossbulat: Addressing prompt move to `ui-overlay` in https://github.com/polkadot-cloud/polkadot-staking-dashboard/pull/3555

PR #3567, opened 2026-06-15
perf(bundle): lazy-load page routes and split heavy vendor chunks

joeljkrb:
## Summary

All 8 pages were static imports, so heavy page-only dependencies (chart.js,
chartjs-plugin-annotation, chroma-js, etc.) landed in the entry chunk even
though modals and canvases were already lazy-loaded.

## Changes

- Convert every `PagesConfig` entry to `React.lazy(() => import(...))` so each
  route is emitted as its own chunk and loaded on navigation.
- Wrap the page render in `PageWithTitle` with a `Suspense` boundary using the
  existing `PagePreloader` fallback.
- Widen `PageItem.Entry` to `ComponentType<PageProps>` so lazy components are
  assignable.
- Add `manualChunks` to split `charts` (chart.js & friends), `dedot`, and
  `@fortawesome` into separate cacheable vendor chunks.

### DIFF
--- a/packages/app/src/config.ts
+++ b/packages/app/src/config.ts
@@ -10,15 +10,35 @@ import {
 	faUserGear,
 	faUsers,
 } from '@fortawesome/free-solid-svg-icons'
-import { Nominate } from 'pages/Nominate'
-import { Operators } from 'pages/Operators'
-import { Overview } from 'pages/Overview'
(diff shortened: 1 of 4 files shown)


rossbulat:
duplicate of https://github.com/polkadot-cloud/polkadot-staking-dashboard/pull/3390

PR #3571, opened 2026-06-15
perf(worker): use BigInt compare for staker sort instead of BigNumber

joeljkrb:
## Summary

The stakers worker sorted each validator's `others` by bonded value using a
`new BigNumber(...).minus(...)` per comparison. The values are planck integer
strings, so a native `BigInt` compare does the same job without allocating a
BigNumber object on every comparison across thousands of nominators.

## Changes

- Replace the `BigNumber`-based comparator with `BigInt(b.value) - BigInt(a.value)`.
- Drop the now-unused `bignumber.js` import from the worker.

### DIFF
--- a/packages/workers/src/stakers.ts
+++ b/packages/workers/src/stakers.ts
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 import { planckToUnit } from '@w3ux/utils'
-import BigNumber from 'bignumber.js'
 import type { ActiveAccountOwnStake, Staker } from 'types'
 import type { ProcessExposuresArgs } from './types'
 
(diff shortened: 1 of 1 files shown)


packages/workers/src/stakers.ts
```diff
@@ -35,10 +34,12 @@ const processExposures = (data: ProcessExposuresArgs) => {
 			})) ?? []
 
 		if (others.length) {
-			// Sort `others` by value bonded, largest first.
+			// Sort `others` by value bonded, largest first. Values are planck
+			// integer strings, so a BigInt compare avoids per-comparison BigNumber
+			// allocation.
```
rossbulat:
```suggestion
			// Sort `others` by value bonded, largest first.
```

rossbulat (approved):

PR #3583, opened 2026-06-19
security: localStorage parse guards and safe external links

joeljkrb:
## Summary

Two self-contained commits, no behavioural change for normal users, they close a crash vector and tighten outbound links without altering app logic.

## 1. Guard JSON.parse of localStorage

Corrupt or poisoned localStorage currently throws in the render path (e.g. a malformed `reserve_balances` entry crashes the balances view).

Wraps the raw parses in try/catch with safe fallbacks:
- `useBalances/util.ts` - `reserve_balances` → falls back to defaults
- `EraStakers/util.ts` - `*_exposures` → clears bad key, treats as missing
- `Validators/Utils.ts` - `*_favorites` → `[]`; `*_validators` → clears, treats as missing
- `locales/util/language.ts` - `lng_resources` → falls back to dynamic load

Valid data behaves exactly as before; only corrupt data now degrades gracefully. The two cache-style keys self-heal by clearing the bad entry.

## 2. Harden external links

- Adds `noopener,noreferrer` to all 12 `window.open(..., '_blank')` calls so opened pages can't reach back via `window.opener` and the Referer header isn't leaked.
- Validates the operator `website` field (third-party `@w3ux/validator-assets` data) through `new URL()` and only opens `http(s):` schemes - rejecting `javascript:`/`data:` injection.

Touched: Operators item, header account/menu popovers, help tooltip, quick actions, balance links, connect-item web URL popover.


joeljkrb:
Heads up: the failing `check` job is not caused by this PR - it's the Biome 2.5.0 SVG-parsing breakage in `packages/assets` that landed on `main` with #3575. Fixed separately in #3584. Once that's merged and this branch is rebased, CI will pass.

PR #3594, opened 2026-06-26
feat(locales): add Russian (ru) language support

joeljkrb:
## Summary

Adds **Russian (`ru`)** as a fully supported language for the Polkadot Cloud
Staking dashboard.

## Changes

**Translations (5 new files, ~1,485 lines):**
- `packages/locales/src/resources/ru/app.json`
- `packages/locales/src/resources/ru/help.json`
- `packages/locales/src/resources/ru/modals.json`
- `packages/locales/src/resources/ru/pages.json`
- `packages/locales/src/resources/ru/tips.json`

Each file mirrors the English key structure exactly (validated below).

## Russian pluralisation

Unlike the existing languages, Russian has **three integer plural forms**, so
every English `_one`/`_other` pair was expanded into Russian
`_one`/`_few`/`_many` with correct noun declension:

| count | form | example |
|---|---|---|
| 1, 21, 31… | `_one` | `{{count}} валидатор` |
| 2–4, 22–24… | `_few` | `{{count}} валидатора` |
| 0, 5–20, 25… | `_many` | `{{count}} валидаторов` |

- **24 plural groups** expanded across `app` (12), `modals` (10) and `pages` (2).
- The explicit **`_zero`** form is preserved where English defines a special
  empty-state message (`notification`, `activeRoles`, `importedAccount`).
- The locale validator collapses plural suffixes to the base key, so the key
  structure still matches English exactly.

## Translation conventions

- **Register:** formal «вы» throughout; Russian sentence case (not English
  Title Case).
- **Transliterated where standard in the Russian crypto community:** пул,
  стейкинг, Прокси, рантайм, халвинг, слэшинг, очки эры.
- **Kept in Latin (brand / protocol identifiers):** Polkadot, Kusama, Ledger,
  Polkadot Vault, Staking API, Polkawatch, Relay Chain, the Root / Bouncer /
  Depositor pool roles, and Stash.
- **Consistent glossary** across all files: Заблокировать / Разблокировать
  (bond / unbond), Номинировать (nominate), Реинвестировать (compound),
  Вывести (withdraw), Получить (claim), вознаграждения (rewards),
  выплаты (payouts).


rossbulat:
It looks like we need the dynamic text size support in the side menu:

Screenshot 2026-06-28 at 12 18 37

joeljkrb:
@rossbulat - Long translated nav labels (e.g. Russian "Вознаграждения") wrapped under their icon in the fixed-width side menu. Labels now stay on one line and auto-scale their font down to fit the available width, with ellipsis as a fallback at the minimum size. Adds a reusable `useFitText` hook that measures overflow and re-fits on text or container changes.

packages/hooks/src/useFitText/index.ts
rossbulat, line 1: @joeljkrb Can we move this hook to a separate PR?

Is it similar to this?: https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/main/packages/hooks/src/useInputAutoFontSize/index.ts

If so, there might be functionality they can both re-use to prevent duplication.

joeljkrb: Yes, happy to split it out. It's similar in goal but not mechanics: useInputAutoFontSize needs a hidden mirror span because inputs don't expose their natural text width, while useFitText just compares scrollWidth/clientWidth on a static element and needs none of that. 

The only truly shared code is the one-line proportional scale formula, so I'd keep them as separate hooks.

PR #3612, opened 2026-07-01
feat(fix): Trim whitespace from pasted addresses in account input

joeljkrb:
Pasting an SS58 address copied from a block explorer or wallet often carries a leading or trailing space. `AccountInput.handleChange` stored the raw value and ran `isValidAddress` on it, so the field showed "Invalid" and the Import button stayed disabled even though the address was correct, with no visible clue that an invisible space was the cause.

### DIFF
--- a/packages/app/src/library/AccountInput/index.tsx
+++ b/packages/app/src/library/AccountInput/index.tsx
@@ -51,8 +51,6 @@ export const AccountInput = ({
 	const [successLock, setSuccessLocked] = useState<boolean>(locked)
 
 	const handleChange = (e: FormEvent<HTMLInputElement>) => {
-		// Trim surrounding whitespace so addresses pasted from an explorer/wallet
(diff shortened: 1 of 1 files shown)


packages/app/src/library/AccountInput/index.tsx
```diff
@@ -51,7 +51,9 @@ export const AccountInput = ({
 	const [successLock, setSuccessLocked] = useState<boolean>(locked)
 
 	const handleChange = (e: FormEvent<HTMLInputElement>) => {
-		const newValue = e.currentTarget.value
+		// Trim surrounding whitespace so addresses pasted from an explorer/wallet
+		// (which often carry a leading/trailing space) still validate and import.
```
rossbulat:
```suggestion
```

rossbulat (approved):

PR #3638, opened 2026-07-07
feat: bootstrap stablecoins app

rossbulat:
## Summary

Adds the user-facing Stablecoins app on top of the shared stablecoin service layer. Users can review their holdings and send supported stablecoins across Polkadot Hub and Hydration with a configurable fee asset.

## Changes

- Add a balances popover with totals by chain, asset, and portfolio mix
- Add the stablecoin send flow with chain, account, token, amount, and fee-asset selection
- Account for transaction fees and existential deposits when calculating the maximum sendable balance
- Detect and configure Hydration’s selected fee currency before submitting a transfer
- Extend transaction submission to support custom fee estimation, fee display metadata, and signer payload options
- Add app-specific i18n profiles and Stablecoins translations for all supported locales
- Add Dedot feature flags so the Stablecoins app can disable staking and nomination-pool subscriptions
- Move shared support and language modals into `ui-modals`
- Hide staking-specific header and menu features in the Stablecoins app

## Validation

- GitHub `verify (24.x)` check passed
- GitHub license check passed
- `git diff --check`


joeljkrb:
Really solid progress on this. The overall shape is good: the adapter-behind-facade service design fits existing conventions, balance math is BigInt end to end, and the tx-submit generalisation holds up well under comparison (Ledger payloads byte-identical, proxy wrapping unchanged). Typecheck, biome, tests, and the staking build all pass on this branch, and the earlier Copilot `useOutsideAlerter` finding is already fixed.

A few things potentially worth addressing:

### Blockers

1. **`app-stablecoins` cannot build or run.** [package.json](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/app-stablecoins/package.json#L65) pins `vite 8.0.14`, which is incompatible with the workspace `rolldown 1.1.4` override, so `vite dev`/`build` crash on startup (`viteWasmFallbackPlugin` import error). Bumping to `8.1.3` (and `vite-plugin-checker` to `^0.14.4`) fixes it. Related: no CI workflow builds this package, which is why CI is green; worth adding its `build:verify` to `ci.yml` in this PR.

2. **HOLLAR reads and writes target different ledgers** ([hydration.ts#L151-L161](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/dedot-api/src/stablecoins/hydration.ts#L151-L161)). The balance is read from the ERC-20 contract via `eth_call`, but the transfer is built as `tokens.transferKeepAlive(recipient, 222, amount)`. Asset 222 is registered as `Erc20` type, and on-chain the tokens pallet holds ~6.5k HOLLAR while the contract supply is ~11.8M, so transfers will fail with `BalanceTooLow`. Routing Hydration transfers through `currencies.transfer` (which routes Erc20 assets via the EVM adapter) should fix it; worth verifying end to end on chopsticks.

3. **The lazy Hydration connect is racy and leaks websockets** ([polkadot.ts#L209-L217](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/dedot-api/src/services/polkadot.ts#L209-L217)). The `!this.apiHydration` check happens before an await, and the first balances load fires 4 concurrent Hydration calls, so multiple `DedotClient`s are created and the extras stay connected forever, untouched by `unsubscribe()`. Memoizing the in-flight promise (`this.hydrationApi ??= DedotClient.new(...)`) resolves it.

4. **A stale tx stays submittable while a new one builds** ([Send.tsx#L474-L512](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/app-stablecoins/src/pages/Send.tsx#L474-L512)). Changing amount/recipient/chain does not clear `transferTx` until the async rebuild resolves (seconds when Hydration lazily connects), and `validTransfer` only checks `!!transferTx`, so clicking Send in that window submits the old values. Clearing `transferTx` synchronously before the await closes the gap.

5. **WalletPopover renders hardcoded placeholder balances** ([WalletPopover.tsx#L14-L64](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/app-stablecoins/src/components/WalletPopover.tsx#L14-L64)). Wiring it to the existing `useStablecoinBalances` hook, or hiding the button until real data lands, would avoid shipping mock data.

### Important (money path)

- Four of the seven hardcoded existential deposits in [consts/stablecoins.ts](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/consts/src/stablecoins.ts#L47) are wrong vs live chain state: AH DOT is 100x too high, AH USDC/USDT 70x too high, and HOLLAR 200x too low (a max send would fail keep-alive). Fetching `minBalance` / registry ED at runtime would prevent drift.
- Pending tx counts are scoped per network ([txSubmit/index.ts#L87-L92](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/global-bus/src/txSubmit/index.ts#L87-L92)), but this app submits to two chains per network, so a pending AH tx inflates the nonce for a Hydration submission from the same account. Keying by `genesisHash` would fix it.
- The new on-chain nonce fetch ([useSubmitExtrinsic/index.tsx#L256-L258](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/tx-submit/src/useSubmitExtrinsic/index.tsx#L256-L258)) sits between `setUidSubmitted(uid, true)` and signing with no try/catch. An RPC failure leaves the button stuck at "Submitting" with no retry possible. This affects staking too.
- While the Hydration fee-currency query is in flight (or rejects), `hydrationFeeCurrency` is null and reads as "no setup needed" ([Send.tsx#L411-L415](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/app-stablecoins/src/pages/Send.tsx#L411-L415)), so a user can submit and pay fees in HDX they do not hold. Treating null as not-ready closes it.
- `maxSendableBalance` ignores `frozen` ([Send.tsx#L120-L134](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/app-stablecoins/src/pages/Send.tsx#L120-L134)) even though both adapters populate it, so frozen funds pass validation and fail on-chain.
- The form boots pre-filled with `1000.00` and the first account as recipient ([Send.tsx#L252](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/app-stablecoins/src/pages/Send.tsx#L252)). Defaulting to empty seems safer for a payments app.
- Failure modes are silent throughout: RPC errors become zero balances, fee estimation errors coerce to `0n` (which also disables the insufficient-funds warning), unreachable Hydration endpoints hang `balances()` forever (WsProvider retries indefinitely) and take AH balances down with it, and the async effects in Send.tsx / `useStablecoinBalances` have no catch. Error states, `Promise.allSettled` per chain, and a connect timeout would cover it.
- The shared header can open `Networks`, `SelectLanguage`, `SyncAccounts`, and `Transfer`, none of which are registered in [Overlays.tsx#L41-L49](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/app-stablecoins/src/Overlays.tsx#L41-L49), so those buttons open empty modals.
- Extension prompts at submit time say "Polkadot Cloud Staking" ([useSubmitExtrinsic/index.tsx#L146](https://github.com/polkadot-cloud/polkadot-cloud-apps/blob/de8dd1d0fbcfa12be158f70c4a7e7a3aa2d0032d/packages/tx-submit/src/useSubmitExtrinsic/index.tsx#L146) uses `StakingDappName`). Making the dapp name injectable fixes it for this app.
- All user-facing copy is hardcoded English while the app wires up i18next; moving strings to `locales` would match convention.

### Housekeeping

- The "update package.json files" commit downgrades deps relative to main (radix-ui, react-fontawesome, react-i18next, i18next, openai, @types/node), and the new package pins older versions still (radix-ui `1.4.2`, react `19.2.6`, vite `8.0.14`). Looks like a stale-branch artifact worth re-syncing.
- A couple of unrelated changes are riding along (the ValidatorMetrics commission display, the `usePageFromHash` barrel export removal). Might be cleaner as separate PRs.
- Smaller items: `og-image.png` is byte-identical to staking's; the inline Buffer polyfill added to both `index.html` files has stray indentation; some dead SCSS selectors and unused deps in the new package; and the 15-line stablecoins stub is copy-pasted into four services (could be a shared `defaultStablecoinsInterface`).

The chain-level claims (asset types, EDs, storage state) were verified against live Asset Hub and Hydration RPC.

PR #3670, opened 2026-07-20
chore(locales): detect orphaned keys and harden locale validator

joeljkrb:
## Summary
The locale key validator (`packages/locales/.scripts/localeValidate.cjs`) had three problems:

1. **One-directional.** It only flagged keys *missing* from a translation (in `en`, absent elsewhere). Keys *orphaned* in a translation, present there but no longer in `en`, e.g. left behind when an English key is removed, were never detected, so dead translation entries accumulated silently.
2. **A guard that could mask missing keys.** The missing-key check only ran when `enKeys.length !== otherKeys.length`. A locale with one missing key *and* one orphaned key has an equal count, so the check was skipped entirely.
3. **Fragile control flow.** It used async `fs.readdir` with `throw` inside the callback (surfaces as an unhandled rejection, not a clean failure), and fired both validation passes concurrently with nondeterministic ordering.

## Changes
- **Bidirectional parity** reports both missing and orphaned keys (set difference both ways), grouped by locale + file.
- **Synchronous & deterministic** switched to `fs.readdirSync`/`readFileSync`; a top-level `try/catch` prints one aggregated report and exits non-zero via `process.exit(1)`.
- **Aggregated reporting** collects every problem across all files/locales instead of throwing on the first.
- **Removed the length-equality guard** so a compensating orphan can no longer hide a missing key.
- **`ORPHANED_KEYS_FATAL` flag (default `false`)** orphaned keys are reported as warnings for now so CI stays green while the pre-existing ones are cleaned up; flip to `true` to enforce strict parity going forward.
- Added missing-file detection and a success message.

## Pre-existing orphans surfaced (as warnings)
The validator now reports **17 orphaned keys** that had already drifted in:

| Locale | File(s) | Count |
|---|---|---|
| `ko` | app.json, pages.json | 3 |
| `pt` | app.json, pages.json | 3 |
| `zh` | help.json | 11 |

The `ko`/`pt` ones are dead strings (`app.documentation`, `pages.alreadyInPool`, `pages.alreadyNominating`), safe to delete. The `zh` set includes a fully-removed `stashAndControllerAccounts` help definition plus some array-index divergences that need a translator's eye. Cleaning these up and flipping `ORPHANED_KEYS_FATAL` to `true` is left as a follow-up.


joeljkrb:
## Orphaned key cleanup (17 keys)
- **`ko` / `pt`** removed dead strings `app.documentation`, `pages.alreadyInPool`, `pages.alreadyNominating` (no longer in `en`).
- **`zh/help.json`**
  - Removed the `stashAndControllerAccounts` definition, fully removed from `en` (8 dead keys).
  - `decentralizationAnalyticsPeriod`,`zh` had split `en`'s 2 paragraphs into 4; merged back to 2 to match `en` (**no content lost**).
  - `activeNominators` dropped a `zh`-only extra sentence so it mirrors `en` exactly.

